PR #6, opened 2021-06-03
add order task done

mankokolya:


src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,19 @@
...
+
+        <mapping class="mate.academy.model.Movie"/>
+        <mapping class="mate.academy.model.CinemaHall"/>
+        <mapping class="mate.academy.model.MovieSession"/>
+        <mapping class="mate.academy.model.User"/>
+        <mapping class="mate.academy.model.Ticket"/>
+        <mapping class="mate.academy.model.ShoppingCart"/>
+        <mapping class="mate.academy.model.Order"/>
+    </session-factory>
+</hibernate-configuration>
```
dant0n, line 19: Empty line

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,69 @@
...
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @OneToMany
+    @JoinColumn(name = "order_id")
+    private List<Ticket> tickets;
+    private LocalDateTime dateTime;
+
+    @ManyToOne
+    private User user;
```
dant0n, line 27: Probably need to fetch lazy as well, and fetch them in orderDao method

dant0n (changes requested):

dant0n:
Got answer in slack, pre-approving

dant0n (approved):

src/main/java/mate/academy/model/Order.java
mankokolya: I made it on purpose as Order closely connected to User and it makes sense to load User with the order at once and User doesn't have any custom fields so will not load too much unnecessary data .

MaliukDaria (changes requested):
Travis failed.... 
I looked a little at the error with which it crashes, perhaps this is due to spaces in the line after serverTimezone and "="...  remove then and try again  <property name="connection.url">jdbc:mysql://localhost:3306/ticket_app?serverTimezone=UTC</property>

MaliukDaria (changes requested):

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -23,12 +27,17 @@ public User login(String email, String password) throws AuthenticationException
...
-        return user;
+    public User register(String email, String password) throws AuthenticationException {
+        try {
+            User user = new User();
+            user.setEmail(email);
+            user.setPassword(password);
+            userService.add(user);
+            shoppingCartService.registerNewShoppingCart(user);
+            return user;
+        } catch (ConstraintViolationException e) {
```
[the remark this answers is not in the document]
mankokolya: In this case we make one request to db but if we will use findByEmail in the worst case we will make two requests.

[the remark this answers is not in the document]
MaliukDaria: Agree

src/main/java/mate/academy/security/AuthenticationServiceImpl.java
```diff
@@ -23,12 +28,17 @@ public User login(String email, String password) throws AuthenticationException
     }
 
     @Override
-    public User register(String email, String password) {
-        User user = new User();
-        user.setEmail(email);
-        user.setPassword(password);
-        userService.add(user);
-        return user;
+    public User register(String email, String password) throws AuthenticationException {
```
MaliukDaria, line 31:
```suggestion
    public User register(String email, String password) throws RegistrationException {
```

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,58 @@
...
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public List<Order> getOrdersHistory(User user) {
+        try (Session session = sessionFactory.openSession()) {
+            return session.createQuery("from Order o join fetch o.tickets t "
+                    + "left join fetch t.user "
```
MaliukDaria, line 48: Do we need to fetch user here?

MaliukDaria (changes requested):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
mankokolya: Yes because I use it in Order toString() method and it is marked with fetch type lazy

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #8, opened 2021-06-08
added solution

safordog:


src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,48 @@
...
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public List<Order> getAllByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("FROM Order o LEFT JOIN FETCH o.tickets "
+                    + "LEFT JOIN FETCH o.user "
```
mankokolya, line 41: no need to fetch user here as you have default fetch type eager in your entity

mankokolya: You have to correct ticket class and use moviesession with fetch type eager instead of Movie and cinemaHall. I asked mentors about this.

mankokolya (changes requested):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
safordog: Дякую за перевірку! Переробив.

mankokolya (approved):

MaliukDaria (approved):

PR #19, opened 2021-06-10
added solution

KseniaObertii:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    private List<Ticket> ticketList;
```
vuyko23, line 20: Lets use cross referance here

vuyko23 (approved):

src/main/java/mate/academy/model/Order.java
KseniaObertii: done

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,74 @@
...
+    @JoinTable(name = "orders_tickets",
+        joinColumns = @JoinColumn(name = "order_id"),
+        inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    private List<Ticket> ticketList;
+    private LocalDateTime orderDate;
+    @ManyToOne
+    private User user;
+
+    public Order() {
+    }
```
MaliukDaria, line 32: Do we need this?

MaliukDaria (approved):

PR #25, opened 2021-06-10
first solution.

VladislavTetar:


Dalai-S (approved):

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setUser(shoppingCart.getUser());
+        order.setTickets(new ArrayList<>(shoppingCart.getTickets()));
+        order.setOrderDate(LocalDateTime.now());
```
MaliukDaria, line 24: Let's clear shopping cart after completing the order

MaliukDaria (changes requested):

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
VladislavTetar: done

PR #27, opened 2021-06-10
All done

Nazar-yo:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.persistence.OneToMany;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
```
Trident1998, line 19: if we write fetch type LAZY, it will good practice. Also let's change the name of our cross reference table and it's columns

Trident1998 (changes requested):

Trident1998 (approved):
LGTM!

Resci (approved):

PR #32, opened 2021-06-10
Done solution order

Elia2909:


src/main/java/mate/academy/model/User.java
```diff
@@ -49,4 +50,14 @@ public void setPassword(String password) {
...
         this.salt = salt;
     }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + "id=" + id
+                + ", email='" + email + '\''
+                + ", password='" + password + '\''
+                + ", salt=" + Arrays.toString(salt)
```
MaliukDaria, line 60: It's not safe to add password and salt fields in toString() method. Lets remove them

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #76, opened 2021-09-14
Initial solution

VitaliyJvM:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,77 @@
...
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @OneToMany
+    private List<Ticket> tickets;
+    private LocalDateTime orderDate;
+
+    @OneToOne
```
RomanDubovskyi, line 26: let's use here lazy strategy

RomanDubovskyi (approved):

PR #78, opened 2021-10-01
implemented order

kovalskiy7777:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,64 @@
...
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    private List<Ticket> tickets;
+    private LocalDateTime orderDate;
+    @ManyToOne
```
AlexRozv, line 22: good practice not to use eager fetch type

AlexRozv (approved):
well done

src/main/java/mate/academy/model/Order.java
kovalskiy7777: Thanks, fixed

Resci (approved):

PR #120, opened 2021-11-03
order was implemented

StasShepet:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,66 @@
...
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    private List<Ticket> tickets;
+    private LocalDateTime orderDate;
+    @ManyToOne
+    private User user;
+
```
taras-oleksiuk, line 24: Let's produce an empty constructor. We will follow the Proxy design pattern

taras-oleksiuk (changes requested):

taras-oleksiuk (approved):

Iryna-U (approved):

PR #143, opened 2022-01-02
hibernate-order-hw solution

Savujkee:


src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +53,36 @@ public static void main(String[] args) {
...
 
         System.out.println(movieSessionService.get(yesterdayMovieSession.getId()));
         System.out.println(movieSessionService.findAvailableSessions(
-                        fastAndFurious.getId(), LocalDate.now()));
+                fastAndFurious.getId(), LocalDate.now()));
+
+        AuthenticationService authenticationService =
+                (AuthenticationService) injector.getInstance(AuthenticationService.class);
+        User elon = null;
+        authenticationService.register("elon.musk@spacex.com", "MyBigRocket");
```
Panchenko-Volodymyr, line 68: we use login and password two times so we can separate it like 
String login = "elon.musk@spacex.com";
and do the same with password

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Override
+    public List<Order> findByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("FROM Order o "
+                    + "LEFT JOIN FETCH o.tickets t "
+                    + "LEFT JOIN FETCH t.movieSession ms "
+                    + "LEFT JOIN FETCH ms.movie "
+                    + "LEFT JOIN FETCH ms.cinemaHall "
+                    + "WHERE o.user =:user", Order.class);
```
Panchenko-Volodymyr, line 46:
```suggestion
            Query<Order> query = session.createQuery("SELECT DISTINCT o FROM Order o "
                    + "LEFT JOIN FETCH o.tickets t "
                    + "LEFT JOIN FETCH o.user "
                    + "LEFT JOIN FETCH t.movieSession ms "
                    + "LEFT JOIN FETCH ms.movie "
                    + "LEFT JOIN FETCH ms.cinemaHall "
                    + "WHERE o.user =:user", Order.class);
```

Panchenko-Volodymyr (changes requested):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
Savujkee: I've tested this before - works just fine without fetching user.
Probably because we pass the user as a parameter.

Panchenko-Volodymyr (approved):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
KseniiaMakarova, line 46: - are we interested in orders without tickets, tickets without moviesession, and so on?
- how many orders will this query return if you have `n` orders with `n` tickets in each order?

KseniiaMakarova: @Savujkee 
> I've tested this before - works just fine without fetching user.
> Probably because we pass the user as a parameter.

not really, check your default fetch type for user in order

KseniiaMakarova (changes requested):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
Savujkee: apparently, n squared
gotcha

KseniiaMakarova (approved):

PR #148, opened 2022-01-03
implemented task hibernate-order

Nata-Vidzivashets:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,71 @@
...
+import javax.persistence.OneToMany;
+
+@Entity(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    @JoinTable(joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
```
ArtikVN, line 23: you can add the correct table name

ArtikVN (approved):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+    }
+
+    @Override
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> getOrdersByUserQuery = session.createQuery("FROM orders o "
+                    + "JOIN FETCH o.user u "
+                    + "LEFT JOIN FETCH o.tickets t "
```
KseniiaMakarova, line 43: - are we interested in orders without tickets, tickets without moviesession, and so on?
- how many orders will this query return if you have `n` orders with `n` tickets in each order?

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #149, opened 2022-01-03
Order model with dao and service layers was implemented.

farfardesu:


src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +55,45 @@ public static void main(String[] args) {
...
+            authenticationService.register("newuserTwo@gmail.com", "123qwe");
+        } catch (Exception e) {
+            throw new RuntimeException("Can't register user.");
+        }
+        try {
+            testUserOne = authenticationService.login("newuserOne@gmail.com", "qwe123");
+            testUserTwo = authenticationService.login("newuserTwo@gmail.com", "123qwe");
+        } catch (AuthenticationException e) {
+            throw new RuntimeException("Can't login with given properties.");
+        }
```
PineappleApplePinecone, line 84: I think, in this project there is no point in this try catch blocks, you can just register new User

PineappleApplePinecone (approved):

src/main/java/mate/academy/Main.java
farfardesu: If try catch blocks will be removed, methods login() an register() may throw an Exception, so in terms of good practice we should catch exceptions to prevent adding `throws SomeException`  in main signature

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Override
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("SELECT DISTINCT o FROM Order o "
+                    + "LEFT JOIN FETCH o.tickets t "
+                    + "LEFT JOIN FETCH o.user u "
+                    + "LEFT JOIN FETCH t.movieSession ms "
+                    + "LEFT JOIN FETCH ms.cinemaHall "
+                    + "LEFT JOIN FETCH ms.movie "
```
KseniiaMakarova, line 46: - are we interested in orders without tickets, tickets without moviesession, and so on?

KseniiaMakarova (approved):

PR #154, opened 2022-01-03
create Order model, dao and service layer

kateryna-deshchinska:


src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +54,35 @@ public static void main(String[] args) {
...
+                (AuthenticationService) injector.getInstance(AuthenticationService.class);
+        try {
+            authenticationService.register("katedeshchynska@gmail.com", "12345678");
+            System.out.println(authenticationService
+                    .login("katedeshchynska@gmail.com", "12345678"));
+        } catch (AuthenticationException e) {
+            throw new RuntimeException(e);
+        }
+        UserService userService = (UserService) injector.getInstance(UserService.class);
+        User user = userService.findByEmail("katedeshchynska@gmail.com").get();
```
ChmokiMoki, line 76: AuthenticationService.login() return User. Maybe be better to use AuthenticationService to find a user instead of UserService

ChmokiMoki (changes requested):

ChmokiMoki (approved):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> orderHistoryQuery = session.createQuery("select distinct o from Order o "
+                    + "left join fetch o.tickets t "
+                    + "left join fetch o.user "
+                    + "left join fetch t.movieSession ms "
+                    + "left join fetch ms.movie "
+                    + "left join fetch ms.cinemaHall "
+                    + "where o.user =:user");
```
KseniiaMakarova, line 47: - are we interested in orders without tickets, tickets without moviesession, and so on?

KseniiaMakarova (approved):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
kateryna-deshchinska: we aren't interested, I agree)

kateryna-deshchinska: we need to use inner join, instead left

PR #169, opened 2022-01-15
add Order, OrderDao, OrderService

annaRav:


src/main/java/mate/academy/dao/OrderDao.java
```diff
@@ -0,0 +1,11 @@
+package mate.academy.dao;
+
+import java.util.List;
+import mate.academy.model.Order;
+import mate.academy.model.User;
+
+public interface OrderDao {
+    List<Order> getOrdersByUser(User user);
+
+    Order getUnfinishedOrderByUser(User user);
```
MaliukDaria, line 10: Why do we need this method? User could have only one unfinished order?

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public Order getUnfinishedOrderByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            CriteriaBuilder criteriaBuilder = session.getCriteriaBuilder();
+            CriteriaQuery<Order> criteriaQuery = criteriaBuilder.createQuery(Order.class);
+            Root<Order> root = criteriaQuery.from(Order.class);
+            Predicate userPredicate = criteriaBuilder.equal(root.get("user"), user);
+            Predicate unfinishedPredicate = criteriaBuilder.equal(root.get("orderDate"), null);
```
MaliukDaria, line 40: It seems to me that it is not a good idea to determine whether an order has been completed or not by the null orderDate field. It's better to add some status column in orders table, where statuses could be FINISHED, UNFINISFED, ... But I don't think we need this in this task

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+
+    @Override
+    @Transactional
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = orderDao.getUnfinishedOrderByUser(shoppingCart.getUser());
+        order.setOrderDate(LocalDateTime.now());
+        return order;
```
MaliukDaria, line 24: If we dont have `add(Order order)` method where can orders come from in the database?

MaliukDaria (changes requested):
Do you test this solution? Does it work without hibernate.cfg.xml file?

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+@Service
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+
+    @Override
+    @Transactional
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setTickets(shoppingCart.getTickets());
```
KseniiaMakarova, line 23: please see checklist

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+        order.setOrderDate(LocalDateTime.now());
+        order.setStatus("completed");
```
KseniiaMakarova, line 25: what about user?

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
annaRav: set copy of list

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,81 @@
...
+    @ManyToOne
+    private User user;
+
+    @OneToMany
+    private List<Ticket> tickets = new ArrayList<>();
+
+    @Column(name = "order_date")
+    private LocalDateTime orderDate;
+
+    private String status;
```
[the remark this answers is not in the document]
annaRav: ok

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
annaRav: agree

MaliukDaria (approved):

PR #196, opened 2022-04-14
Hw hibernate order

OlehMytiai:


src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Override
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("from Order o "
+                            + "left join fetch o.tickets t "
+                            + "left join fetch t.movieSession ms "
+                            + "left join fetch ms.movie "
+                            + "left join fetch ms.cinemaHall "
+                            + "where o.user.id = :userId", Order.class)
```
Liubov-Chernohor, line 45: 1. you could use just "join fetch" without "left" here 
2. where is the join to the "user" field? It should be LIE here....

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+import mate.academy.model.ShoppingCart;
+import mate.academy.model.User;
+import mate.academy.service.OrderService;
+import mate.academy.service.ShoppingCartService;
+
+@Service
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+
```
Liubov-Chernohor, line 18: redundant empty string

Liubov-Chernohor (changes requested):

Liubov-Chernohor (approved):

KostyaNikiforov (approved):

PR #198, opened 2022-04-14
was implemented order dao and order service layers

IgorZima90:


antonignatenko (approved):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    public List<Order> getHistoryByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("from Order o"
+                    + " left join fetch o.tickets t "
+                    + " left join fetch o.user u "
+                    + " left join fetch t.movieSession ms"
+                    + " left join fetch ms.movie "
+                    + " left join fetch ms.cinemaHall"
+                    + " where o.user = :user", Order.class);
+            query.setParameter("user", user);
```
KostyaNikiforov, line 48: Do we need a left join fetch everywhere there?

KostyaNikiforov (changes requested):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
IgorZima90: don't need to use left join fetch because we can take
order History without left join

KostyaNikiforov (approved):

PR #199, opened 2022-04-14
Task completed, Order classes implemented

NazariiYatsiuk:


IgorZima90 (approved):
LGTM :fire:

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("select distinct o from Order o "
+                    + "left join fetch o.user u "
+                    + "left join fetch o.tickets t "
+                    + "left join fetch t.movieSession m "
+                    + "left join fetch m.movie mo "
+                    + "left join fetch m.cinemaHall c "
+                    + "where o.user = :user");
+            query.setParameter("user", user);
```
KostyaNikiforov, line 48: Do we need a left join fetch everywhere there?

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+            throw new DataProcessingException("Can`t add order into DB: " + order, e);
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public List<Order> getByUser(User user) {
```
KostyaNikiforov, line 39:
```suggestion
    public List<Order> findByUser(User user) {
```
or
```suggestion
    public List<Order> getHistoryByUser(User user) {
```
Because, if you return a list, no orders from that user can be found

KostyaNikiforov (changes requested):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("select distinct o from Order o "
+                    + "left join fetch o.user u "
+                    + "left join fetch o.tickets t "
+                    + "left join fetch t.movieSession m "
+                    + "left join fetch m.movie mo "
+                    + "left join fetch m.cinemaHall c "
+                    + "where o.user = :user");
```
KostyaNikiforov, line 47: Do we need a left join fetch everywhere there?

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
KostyaNikiforov, line 39:
```suggestion
    public List<Order> getHistoryByUser(User user) {
```
or
```suggestion
    public List<Order> findByUser(User user) {
```
Because, if you return a list, no orders from that user can be found

KostyaNikiforov (changes requested):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
NazariiYatsiuk: no problem, but 
image

KostyaNikiforov (approved):
Omg, I didn't notice that item on the checklist 😶 
But i still think that a method that may return nothing shouldn't be named "get"
We'll figure it out)

PR #214, opened 2022-05-22
 hibernate-order-hw, could you check it, please

lutskvadym:


src/main/resources/init_db.sql
```diff
@@ -0,0 +1 @@
+CREATE SCHEMA `cinema5` DEFAULT CHARACTER SET utf8;
```
Nick97-git, line 1: do you need this file in this task?

Nick97-git (changes requested):

src/main/resources/init_db.sql
lutskvadym: actually no, thank you, it was deleted

Nick97-git (approved):

PR #232, opened 2022-06-07
Added Order entity. Implement OrderDao and OrderService.

VovaKlimyuk:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,76 @@
...
+    private Long id;
+    @OneToMany
+    @JoinTable(name = "orders_tickets",
+            joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    private List<Ticket> tickets;
+    @Column(name = "order_date")
+    private LocalDateTime orderDate;
+    @OneToOne
+    @MapsId
```
lexzane, line 28: Can `user` have multiple orders?
And are you sure that here we can use annotation `MapsId`? Will everything be ok if user places the second order?

lexzane (changes requested):

KseniiaMakarova (approved):

PR #256, opened 2022-07-24
Created entity Order

ChumakAndrey:
Created and implemented OrderDao, OrderService
Added hibernate.cfg.xml

### DIFF
--- a/src/main/java/mate/academy/Main.java
+++ b/src/main/java/mate/academy/Main.java
@@ -2,16 +2,25 @@ package mate.academy;
 
 import java.time.LocalDate;
 import java.time.LocalDateTime;
+import mate.academy.exception.AuthenticationException;
+import mate.academy.lib.Injector;
 import mate.academy.model.CinemaHall;
 import mate.academy.model.Movie;
 import mate.academy.model.MovieSession;
+import mate.academy.model.ShoppingCart;
+import mate.academy.model.User;
+import mate.academy.security.AuthenticationService;
 import mate.academy.service.CinemaHallService;
 import mate.academy.service.MovieService;
 import mate.academy.service.MovieSessionService;
+import mate.academy.service.OrderService;
+import mate.academy.service.ShoppingCartService;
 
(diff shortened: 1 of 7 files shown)


src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +54,36 @@ public static void main(String[] args) {
...
+                fastAndFurious.getId(), LocalDate.now()));
+
+        User bob;
+        AuthenticationService authenticationService =
+                (AuthenticationService) injector.getInstance(AuthenticationService.class);
+        try {
+            bob = authenticationService.login("bob@test", "1234");
+        } catch (AuthenticationException e) {
+            throw new RuntimeException("Can't register user", e);
+        }
```
antonEEmus, line 73: This is not critical, but you don't need to catch exceptions in main method, for testing, you can just add the exception to method signature using "throws".

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        List<Ticket> tickets = new ArrayList<>(shoppingCart.getTickets());
+        User cartUser = shoppingCart.getUser();
+        Order order = new Order();
+        order.setTickets(tickets);
+        order.setUser(cartUser);
+        order.setOrderDate(LocalDateTime.now());
+        orderDao.add(order);
+        shoppingCartService.clearShoppingCart(shoppingCart);
+        return order;
```
antonEEmus, line 33:
```suggestion
        shoppingCartService.clearShoppingCart(shoppingCart);
        return orderDao.add(order);
```

antonEEmus (changes requested):

antonEEmus (commented):
🔥

antonEEmus (approved):
🔥

KostyaNikiforov (approved):

PR #260, opened 2022-07-26
added solution for hibernate-order-hw task

dmnichz:


src/main/java/mate/academy/dao/OrderDao.java
```diff
@@ -0,0 +1,13 @@
+package mate.academy.dao;
+
+import java.util.List;
+import mate.academy.model.Order;
+import mate.academy.model.User;
+
+public interface OrderDao {
+    Order add(Order order);
+
+    Order getByUser(User user);
```
MaliukDaria, line 10: Why do we need this method? User could have only one Order?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #261, opened 2022-07-26
Hibernate order hw

Cosmic-M:


src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,56 @@
...
+            session = HibernateUtil.getSessionFactory().openSession();
+            transaction = session.beginTransaction();
+            session.save(order);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't insert an order "
+                    + order + " into DB: " + order, e);
```
ChumakAndrey, line 31: I think it redundant to duplicate order in message

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,56 @@
...
+    public Optional<List<Order>> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("FROM Order o "
+                    + "LEFT JOIN FETCH o.user u "
+                    + "LEFT JOIN FETCH o.tickets t "
+                    + "LEFT JOIN FETCH t.movieSession ms "
+                    + "LEFT JOIN FETCH ms.movie "
+                    + "LEFT JOIN FETCH ms.cinemaHall "
+                    + "WHERE u.id = :user_id", Order.class);
+            query.setParameter("user_id", user.getId());
```
ChumakAndrey, line 50: It works and can be as it is, but you can pass to parameter not only user.id but user

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setOrderDate(LocalDateTime.now());
+        order.setUser(shoppingCart.getUser());
+        List<Ticket> tickets = new ArrayList<>(shoppingCart.getTickets());
+        order.setTickets(tickets);
+        orderDao.add(order);
+        shoppingCartService.clearShoppingCart(shoppingCart);
+        return null;
```
ChumakAndrey, line 36: Why do you return null here?

ChumakAndrey (changes requested):

ChumakAndrey (approved):

KostyaNikiforov (approved):

PR #277, opened 2022-08-22
hibernate-order-hw

morozovyrv:


MaliukDaria (approved):
The task should not be implemented in the master/main branch, a separate branch should be created for this (you can leave this task as it is, but create separate branches for next onews 🙂)

PR #299, opened 2022-09-29
HW hibernate-order is done

Valerii-Sh:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,72 @@
...
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    @JoinTable(name = "orders_tickets",
+            joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    private List<Ticket> tickets;
+    private LocalDateTime orderDate;
```
kateryna-deshchinska, line 27:
```suggestion
    @Column(name = "order_date")
    private LocalDateTime orderDate;
```

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setTickets(shoppingCart.getTickets());
```
kateryna-deshchinska, line 24: When passing tickets from shoppingCart to order — pass the copy of this list(you can use ArrayList constructor for this); The reason is that Hibernate will automatically delete these tickets from shoppingCart(in DB) after noticing them in our order object. Let’s do this manually using clear() method from ShoppingCartService after the order with tickets has been created.

kateryna-deshchinska (changes requested):

kozhukhovsky (approved):

PR #303, opened 2022-10-04
implements new entity Order

smshk1:


src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+            session = HibernateUtil.getSessionFactory().openSession();
+            transaction = session.beginTransaction();
+            session.save(order);
+            transaction.commit();
+            return order;
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Couldn't create a order dao: " + order, e);
```
GolDenik, line 29:
```suggestion
            throw new DataProcessingException("Couldn't create an order: " + order, e);
```

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,72 @@
...
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    @JoinTable(name = "orders_tickets",
+            joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    private List<Ticket> tickets;
+    private LocalDateTime orderTime;
```
GolDenik, line 27: You should have good naming in your DB , because without this annotation you will get such a column name as "orderDate".

```suggestion
    @Column(name = "order_date")
    private LocalDateTime orderTime;
```

GolDenik (changes requested):

GolDenik (approved):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+            throw new DataProcessingException("Couldn't create an order: " + order, e);
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public List<Order> getAllByUser(User user) {
```
Noct2000, line 38: it would be better to rename it 
```suggestion
    public List<Order> getOrdersHistory(User user) {
```

Noct2000 (approved):

PR #306, opened 2022-10-05
Order realise

Alexhqolma:


ArturZhadan (approved):

src/main/java/mate/academy/dao/OrderDao.java
```diff
@@ -0,0 +1,11 @@
+package mate.academy.dao;
+
+import java.util.List;
+import mate.academy.model.Order;
+import mate.academy.model.User;
+
+public interface OrderDao {
+    Order add(Order order);
+
+    List<Order> getByUser(User user);
```
Noct2000, line 10: it would be better to rename it 
```suggestion
    List<Order> getOrdersHistory(User user);
```

Noct2000 (approved):

PR #309, opened 2022-10-05
created order service and

Vladyslav-Kondiievskyi:


README.md
```diff
@@ -14,9 +14,9 @@
         }
         ```    
 
-Attention: you can have some problem with naming of the table. Let's use `orders` instead of `order`.
+Attention: you can have some problem with naming of the table. Let's use `orders` instead of `orders`.
 
-__You can check yourself using this__ [checklist](https://mate-academy.github.io/jv-program-common-mistakes/hibernate/add-order/add-order-hw)  
+__You can check yourself using this__ [checklist](https://mate-academy.github.io/jv-program-common-mistakes/hibernate/add-orders/add-orders-hw)
```
kateryna-deshchinska, line 19: Don't change readme file

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,72 @@
...
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    @JoinTable(name = "orders_tickets",
+            joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    private List<Ticket> tickets;
+    private LocalDateTime orderDate;
```
kateryna-deshchinska, line 27:
```suggestion
@Column(name = "order_date")
    private LocalDateTime orderDate;
```

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setTickets(shoppingCart.getTickets());
```
kateryna-deshchinska, line 24: When passing tickets from shoppingCart to order — pass the copy of this list(you can use ArrayList constructor for this); The reason is that Hibernate will automatically delete these tickets from shoppingCart(in DB) after noticing them in our order object. Let’s do this manually using clear() method from ShoppingCartService after the order with tickets has been created.

kateryna-deshchinska (changes requested):

PR #312, opened 2022-10-05
implemented OrderService and it's entity class with dao, couldn't avoid n+1 problem

X3NON-135:


src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,52 @@
...
+import mate.academy.model.Order;
+import mate.academy.model.User;
+import mate.academy.util.HibernateUtil;
+import org.hibernate.Session;
+import org.hibernate.Transaction;
+
+@Dao
+public class OrderDaoImpl implements OrderDao {
+    @Override
+    public Order completeOrder(Order order) {
```
max-bolotin, line 16: The checklist says:
Don’t create method completeOrder(ShoppingCart shoppingCart) on the dao layer. Let’s create add(Order order) instead.

So better call this method add(Order order)

max-bolotin (approved):

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setTickets(new ArrayList<>(shoppingCart.getTickets()));
+        order.setUser(shoppingCart.getUser());
+        order.setOrderDate(LocalDateTime.now());
```
max-bolotin, line 24: checklist says to clear shopping cart after adding tickets to the order, so use method shoppingCartService.clearShoppingCart(shoppingCart);

Noct2000 (approved):

PR #320, opened 2022-10-07
Add: implementation to hibernate order homework

LubusUzh:


Oleksandr-Tovstyi (approved):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Override
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("FROM Order o "
+                    + "LEFT JOIN FETCH o.tickets t "
+                    + "LEFT JOIN FETCH o.user u "
+                    + "LEFT JOIN FETCH t.movieSession ms "
+                    + "LEFT JOIN FETCH ms.movie m "
+                    + "LEFT JOIN FETCH ms.cinemaHall ch "
```
Sofasmile, line 45: why did you use LEFT JOIN here? is it possible to create order without tickets?

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,70 @@
...
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    @JoinTable(name = "orders_tickets",
+                joinColumns = @JoinColumn(name = "order_id"),
+                inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    private List<Ticket> tickets;
+    private LocalDateTime orderDate;
```
Sofasmile, line 27: let's setup name for column: order_date

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setTickets(shoppingCart.getTickets());
```
Sofasmile, line 24: we should use the constructor of ArrayList here - please read common mistakes

Sofasmile (changes requested):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
LubusUzh: I use it to see the message that will print by System.out.println, but in real life, I don't need to use LEFT JOIN for tickets in the Order.
And for the second question - no, it's not possible to create an Order without tickets, so I change the type of joining from LEFT to INNER.

dSkarzh (approved):

Sofasmile (approved):

PR #329, opened 2022-10-09
implemented order service

AshaJaneBrown:


Toltbcm (approved):

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public void setUser(User user) {
+        this.user = user;
+    }
+
+    @Override
+    public String toString() {
+        return "Order{"
+                + "id=" + id
+                + ", tickets=" + tickets
```
dSkarzh, line 69: it is better not to include lazily fetched associations in `toString()`, it can cause `LazyInitializtionException`

dSkarzh (approved):

PR #335, opened 2022-10-09
add solution

AndrewHudzenko:


JK-devv (approved):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,56 @@
...
+            }
+        }
+    }
+
+    @Override
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("from Order o "
+                    + "left join fetch o.user u "
+                    + "left join  fetch o.tickets t "
```
dSkarzh, line 44:
```suggestion
                    + "left join fetch o.tickets t "
```

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public void setUser(User user) {
+        this.user = user;
+    }
+
+    @Override
+    public String toString() {
+        return "Order{"
+                + "id=" + id
+                + ", tickets=" + tickets
```
dSkarzh, line 64: it is better not to include lazily fetched associations in `toString()`, it can cause `LazyInitializtionException`

dSkarzh (approved):

PR #345, opened 2022-10-11
order was created

KosatskiyKirill:


AndrewHudzenko (approved):

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public void setUser(User user) {
+        this.user = user;
+    }
+
+    @Override
+    public String toString() {
+        return "Order{" + "id=" + id
+                + ", tickets=" + tickets
```
dSkarzh, line 69: let's not include lazily fetched fields in `toString()`

dSkarzh (approved):

PR #351, opened 2022-10-24
done, deleted toString from MovieSession not to get lazyInitializatio…

den-superman:
…nException during testing in main


Rommelua (approved):
Good!

PR #363, opened 2022-11-30
Solution was added

Kostiantyn-Sacheiev:


src/main/java/mate/academy/dao/OrderDao.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.List;
+import java.util.Optional;
+import mate.academy.model.Order;
+import mate.academy.model.User;
+
+public interface OrderDao {
+    Order add(Order order);
+
+    Optional<List<Order>> getByUser(User user);
```
aaananas, line 11:
```suggestion
   List<Order> getByUser(User user);
```
dao would return empty list if user doesn't have orders

aaananas (changes requested):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,58 @@
...
+                            + "LEFT JOIN FETCH t.movieSession ms "
+                            + "LEFT JOIN FETCH ms.cinemaHall "
+                            + "LEFT JOIN FETCH ms.movie "
+                            + "WHERE o.user.id = :userId", Order.class)
+                    .setParameter("userId", user.getId())
+                    .getResultList();
+            if (ordersByUser.isEmpty()) {
+                return Collections.emptyList();
+            }
+            return ordersByUser;
```
aaananas, line 53:
```suggestion
            return session.createQuery("SELECT o FROM Order o "
                            + "LEFT JOIN FETCH o.user u "
                            + "LEFT JOIN FETCH o.tickets t "
                            + "LEFT JOIN FETCH t.movieSession ms "
                            + "LEFT JOIN FETCH ms.cinemaHall "
                            + "LEFT JOIN FETCH ms.movie "
                            + "WHERE o.user.id = :userId", Order.class)
                    .setParameter("userId", user.getId())
                    .getResultList();
```

aaananas (approved):

PR #364, opened 2022-11-30
Implement all tasks

CyberDogFK:


src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+    @Inject
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setOrderDate(LocalDateTime.now());
+        order.setUser(shoppingCart.getUser());
```
YuriiHryniv, line 25: I would use here contructor in return statement instead of setters

YuriiHryniv (approved):
LGTM!

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @OneToMany
+    @JoinTable(name = "orders_tickets",
+            joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    private List<Ticket> tickets;
+
+    @JoinColumn(name = "order_date")
+    private LocalDateTime orderDate;
+
```
ChumakAn, line 31: let's remove unnecessary blank spaces between variables

ChumakAn (changes requested):

ChumakAn (approved):

PR #365, opened 2022-11-30
Order is implemented

YuriiHryniv:


src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,33 @@
...
+@Service
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        List<Ticket> tickets = new ArrayList<>(shoppingCart.getTickets());
+        User user = shoppingCart.getUser();
+        shoppingCart.getTickets().clear();
```
CyberDogFK, line 24: We have method in ShoppingCartService as clearShoppingCart(ShoppingCart shoppingCart(), and i think, it's better to use him.
Becouse here you manually removing tickets from memory, not from DB

CyberDogFK (changes requested):

CyberDogFK (approved):

ChumakAn (approved):
Looks good to me 👍

PR #366, opened 2022-12-01
Order solution

arthur4655:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,100 @@
...
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany(fetch = FetchType.LAZY)
+    @JoinTable(name = "orders_tickets",
+            joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    private List<Ticket> tickets;
+    private LocalDateTime orderDate;
```
igorgalenko, line 28: Maybe better to set another name to this field

igorgalenko (approved):
Nice. Just small remark

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,100 @@
...
+import javax.persistence.OneToMany;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany(fetch = FetchType.LAZY)
```
liliia-ponomarenko, line 23: Which default fetch type for relation one-to-many?

liliia-ponomarenko (changes requested):
good job! Let’s fix a few mistakes  ;)

src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +58,52 @@ public static void main(String[] args) {
...
+            try {
+                casualUser = authenticationService.register("user@gmail.com", "userPass1234");
+                registrationSuccess = true;
+            } catch (RegistrationException e) {
+                throw new RuntimeException("Can't complete registration. Try again.", e);
+            }
+        }
+        System.out.println(casualUser);
+        boolean loginSuccess = false;
+        while (!loginSuccess) {
```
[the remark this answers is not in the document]
arthur4655: ohh, I guess there should be system.out.println and then I will get an infinite loop. In general, this is what I wanted, but without the ability to enter new data, it turns out a little wrong)

liliia-ponomarenko (approved):
great 😊

PR #371, opened 2022-12-04
implemented task

vov4ik89:


src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,29 @@
...
+
+@Service
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        return orderDao
+                .add(new Order(shoppingCart.getTickets(), LocalDateTime.now(),
```
igorgalenko, line 21: image

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,86 @@
...
+    @JoinTable(name = "orders_tickets",
+            joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    private List<Ticket> tickets;
+    @Column(name = "order_date")
+    private LocalDateTime orderDate;
+    @OneToOne
+    @MapsId
+    @JoinColumn(name = "id")
+    private User user;
```
igorgalenko, line 34: Not sure about this relation. Can one user have more than one order in the ticket shop?

igorgalenko (changes requested):
I think you have chosen wrong relation between User and Order. Also look to the [checklist](https://mate-academy.github.io/jv-program-common-mistakes/hibernate/add-order/add-order-hw)

liliia-ponomarenko (changes requested):
good job! Let’s  fix a few mistakes ;)

PR #382, opened 2022-12-11
solution

vyshnevsky:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,83 @@
...
+            joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    private List<Ticket> tickets;
+    @Column(name = "order_date")
+    private LocalDateTime orderDate;
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "user_id")
+    private User user;
+
+    public Order() {
```
Veronika-Tretiak, line 34: it is redundant

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+import mate.academy.service.OrderService;
+
+@Service
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        List<Ticket> ticketList = new ArrayList<>(shoppingCart.getTickets());
```
Veronika-Tretiak, line 22: When passing tickets from shoppingCart to order — pass the copy of this list(you can use ArrayList constructor for this); The reason is that Hibernate will automatically delete these tickets from shoppingCart(in DB) after noticing them in our order object. Let’s do this manually using clear() method from ShoppingCartService after the order with tickets has been created.

Veronika-Tretiak (approved):

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public Order(List<Ticket> tickets, LocalDateTime orderDate, User user) {
+        this.tickets = tickets;
+        this.orderDate = orderDate;
+        this.user = user;
+    }
```
Noct2000, line 41: you don't use `Order(List<Ticket> tickets, LocalDateTime orderDate, User user)` constructor, so you can remove these constructors

Noct2000 (changes requested):

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,77 @@
...
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    private List<Ticket> tickets;
+    @Column(name = "order_date")
+    private LocalDateTime orderDate;
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "user_id")
+    private User user;
+
+    public Order() {
+    }
```
Noct2000, line 35: this is redundant

Noct2000 (approved):

PR #384, opened 2022-12-11
added order and order service with impl

apc10bbe:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,67 @@
...
+
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    private List<Ticket> tickets;
+    private LocalDateTime orderDate;
```
GHSpeedWagon, line 22: add table with order_date instead of orderDate

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,67 @@
...
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    private List<Ticket> tickets;
```
GHSpeedWagon, line 21: add column ticket_id instead of now
add column user_id instead of now

GHSpeedWagon (changes requested):
looks good

GHSpeedWagon (approved):

Noct2000 (approved):

PR #393, opened 2022-12-12
Order implementation branch

muzykaVP:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,75 @@
...
+import javax.persistence.OneToMany;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany()
```
hronograf, line 23:
```suggestion
    @OneToMany
```

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+    private OrderDao orderDao;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setTickets(new ArrayList<>(shoppingCart.getTickets()));
+        order.setUser(shoppingCart.getUser());
+        order.setOrderDate(LocalDateTime.now());
+        orderDao.add(order);
+        shoppingCart.getTickets().clear();
```
hronograf, line 26: will this changes be saved to db?
let's use `ShoppingCartService`

hronograf (changes requested):

PR #397, opened 2022-12-12
impl

sequencerr:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,69 @@
...
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @JoinColumn(name = "order_date")
+    private LocalDateTime orderDate;
+    @OneToMany
+    private List<Ticket> tickets;
```
denys-domashevskyi, line 24: Maybe it will be a good idea to use @JoinTable annotation here?

src/main/java/mate/academy/Main.java
```diff
@@ -2,16 +2,26 @@
...
 import mate.academy.service.MovieSessionService;
+import mate.academy.service.OrderService;
+import mate.academy.service.ShoppingCartService;
 
 public class Main {
+    public static final Injector INJECTOR = Injector.getInstance("mate.academy");
+
     public static void main(String[] args) {
-        MovieService movieService = null;
+        MovieService movieService = ((MovieService) INJECTOR.getInstance(MovieService.class));
```
denys-domashevskyi, line 24: Why do you need extra "( )". Check it everywhere

denys-domashevskyi (changes requested):
Nice solution, small suggestions!

denys-domashevskyi (approved):
Good job!)

Noct2000 (changes requested):
add hibernate.cfg.xml

Noct2000 (approved):

PR #424, opened 2023-02-09
HW order

Maks4u:


src/main/java/mate/academy/Main.java
```diff
@@ -2,16 +2,25 @@
...
 import mate.academy.service.MovieSessionService;
+import mate.academy.service.OrderService;
+import mate.academy.service.ShoppingCartService;
 
 public class Main {
-    public static void main(String[] args) {
-        MovieService movieService = null;
+    private static final Injector inject = Injector.getInstance("mate.academy");
+
+    public static void main(String[] args) throws RegistrationException {
```
Uka0001, line 21: I think its better to use try catch instead of throws...

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+            session = HibernateUtil.getSessionFactory().openSession();
+            transaction = session.beginTransaction();
+            session.save(order);
+            transaction.commit();
+            return order;
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't insert an order session: " + order, e);
```
Uka0001, line 30: what do you mean in inserting order session? I think we inserting just order

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("FROM Order o "
+                    + "LEFT JOIN FETCH o.user u "
+                    + "LEFT JOIN FETCH o.tickets t "
+                    + "LEFT JOIN FETCH t.movieSession ms "
+                    + "LEFT JOIN FETCH ms.movie "
+                    + "LEFT JOIN FETCH ms.cinemaHall "
+                    + "WHERE u =:user", Order.class);
```
Uka0001, line 47: I think you have to get user by user id not by user object

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+    @Inject
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setOrderDate(LocalDateTime.now());
+        order.setTickets(new ArrayList<>(shoppingCart.getTickets()));
```
Uka0001, line 26: shopping cart getTickets() method already return List<Tickets>, why do you setTickets new ArrayList instead of shoppingCart.getTickets()?

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,19 @@
...
+
+        <mapping class="mate.academy.model.Movie"/>
+        <mapping class="mate.academy.model.MovieSession"/>
+        <mapping class="mate.academy.model.CinemaHall"/>
+        <mapping class="mate.academy.model.User"/>
+        <mapping class="mate.academy.model.Ticket"/>
+        <mapping class="mate.academy.model.ShoppingCart"/>
+        <mapping class="mate.academy.model.Order"/>
+    </session-factory>
+</hibernate-configuration>
```
Uka0001, line 19: add empty line in the end, Kostiya commented the same issue in my work today :)

Uka0001 (changes requested):

src/main/java/mate/academy/Main.java
Maks4u: idk, no reason what use here l guess, last my work was approved with that realization

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
Maks4u: why? they are both unique

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
Maks4u: image

src/main/resources/hibernate.cfg.xml
Maks4u: image
already have, idk what git want from me

Uka0001 (approved):

src/main/resources/hibernate.cfg.xml
KostyaNikiforov: Here everything is okay

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
KostyaNikiforov: No difference. People use both options

src/main/java/mate/academy/Main.java
KostyaNikiforov: Usually yes, but this is a testing method so if you don't need to process this exception let's leave it like this

KostyaNikiforov (approved):

PR #462, opened 2023-03-12
First commit.

Taurusden:
1. Hibernate.cfg.xml was created and configured.
2. Order clas was created.
3. All methods from Dao and Service layers were implemented.
4. All methods from OrderService were successfully tested in Main.

### DIFF
--- a/src/main/java/mate/academy/model/Ticket.java
+++ b/src/main/java/mate/academy/model/Ticket.java
@@ -11,7 +11,7 @@ import javax.persistence.Table;
 
 @Entity
 @Table(name = "tickets")
-public class Ticket {
+public class Ticket implements Cloneable {
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
@@ -54,4 +54,12 @@ public class Ticket {
             + '}';
     }
 
+    @Override
+    public Ticket clone() {
+        try {
+            return (Ticket) super.clone();
+        } catch (CloneNotSupportedException e) {
+            throw new RuntimeException("Couldn't create clone of Ticket object", e);
+        }
+    }
 }
--- a/src/main/java/mate/academy/service/impl/OrderServiceImpl.java
+++ b/src/main/java/mate/academy/service/impl/OrderServiceImpl.java
@@ -1,0 +1,37 @@
+package mate.academy.service.impl;
+
+import java.time.LocalDateTime;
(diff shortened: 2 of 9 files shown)


src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        List<Ticket> listOfTickets = new ArrayList<>();
+        for (Ticket ticket : shoppingCart.getTickets()) {
+            listOfTickets.add(ticket.clone());
+        }
+        Order order = new Order();
+        order.setOrderDate(LocalDateTime.now());
+        order.setUser(shoppingCart.getUser());
+        order.setTickets(listOfTickets);
+        return orderDao.add(order);
```
olekskov, line 30: please use shopping cart clear method from respective service

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        List<Ticket> listOfTickets = new ArrayList<>();
+        for (Ticket ticket : shoppingCart.getTickets()) {
+            listOfTickets.add(ticket.clone());
+        }
```
olekskov, line 25: Cloning is overkill solution. We just need to use another list object but not all tickets inside
```suggestion
        List<Ticket> listOfTickets = new ArrayList<>(shoppingCart.getTickets());
```

src/main/java/mate/academy/model/Ticket.java
```diff
@@ -54,4 +54,12 @@ public String toString() {
...
     }
 
+    @Override
+    public Ticket clone() {
+        try {
+            return (Ticket) super.clone();
+        } catch (CloneNotSupportedException e) {
+            throw new RuntimeException("Couldn't create clone of Ticket object", e);
+        }
+    }
```
olekskov, line 64: excessive, see comment above

olekskov (changes requested):

olekskov (approved):
Nice job!

PR #481, opened 2023-04-06
Add Order model. Add Dao, Service layers

GalubaGSV:


mishok321 (approved):

src/main/java/mate/academy/Main.java
```diff
@@ -2,16 +2,28 @@
...
 import mate.academy.service.MovieSessionService;
+import mate.academy.service.OrderService;
+import mate.academy.service.ShoppingCartService;
 
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.academy");
+
     public static void main(String[] args) {
-        MovieService movieService = null;
+        MovieService movieService = (MovieService) injector.getInstance(MovieService.class);
```
karma-o, line 26: do not declare services in a method

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,81 @@
...
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany(fetch = FetchType.LAZY)
+    @JoinTable(name = "order_tickets",
+            joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
```
karma-o, line 25: do we need another table for storing `OneToMany` relations?

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order(new ArrayList<>(shoppingCart.getTickets()),
+                LocalDateTime.now(), shoppingCart.getUser());
+        shoppingCartService.clearShoppingCart(shoppingCart);
+        return orderDao.add(order);
```
karma-o, line 27: let's create an order first, and then clear the cart because we can clear it and then face some error in the process of creating an order, our service method is not transactional yet so we will lose data

karma-o (changes requested):

PR #489, opened 2023-04-16
hibernate_order_hw_done

crowlynx:
created hibernate cfg xml file;
created order model;
created order dao plus it's impl;
created order service plus it's impl;
tested all methods from order service in the main method


andriy-shulepa (approved):
Good job!

PR #493, opened 2023-04-18
implement orderdao and orderservice

Nikitos787:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,90 @@
...
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany(cascade = CascadeType.ALL, orphanRemoval = true)
+    @JoinColumn(name = "order_id")
```
IlkoGood, line 24: What will be the column name for tickets id? Have a look in workbench.

src/main/java/mate/academy/model/ShoppingCart.java
```diff
@@ -15,10 +15,8 @@
 public class ShoppingCart {
     @Id
     private Long id;
-    @OneToMany
-    @JoinTable(name = "shopping_carts_tickets",
-            joinColumns = @JoinColumn(name = "shopping_cart_id"),
-            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    @OneToMany(cascade = CascadeType.ALL, orphanRemoval = true)
+    @JoinColumn(name = "shopping_cart_id")
```
IlkoGood, line 19: same, see why `inverseJoinColumns = @JoinColumn(name = "ticket_id"` was here by default

IlkoGood (approved):
Well done!

src/main/java/mate/academy/model/Order.java
Andriy-dz: agree with @IlkoGood . And add name for new columns. And for columns tickets, orderDate. I also recommend removing cascades

src/main/java/mate/academy/model/ShoppingCart.java
Andriy-dz: why did you remove the correct code?)

Andriy-dz (changes requested):

src/main/java/mate/academy/model/ShoppingCart.java
Nikitos787: yes i take too much on myself

Andriy-dz (approved):
Good Solution!

PR #495, opened 2023-04-19
Solution

kateryna-mykh:


src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setOrderDate(LocalDateTime.now());
+        order.setUser(shoppingCart.getUser());
+        order.setTickets(List.copyOf(shoppingCart.getTickets()));
+        shoppingCartService.clearShoppingCart(shoppingCart);
+        return orderDao.add(order);
```
uliana-andrushchuk, line 28: Please, first create an order and then clear the cart. If we clear the cart before creating the order, we may encounter errors in the process.

uliana-andrushchuk (changes requested):

PR #520, opened 2023-04-23
implemented methods

AndrewkeY-hub:


KyslyiViacheslav (approved):
nice

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "Order{"
+                + "id=" + id
+                + ", tickets=" + tickets
+                + ", orderDate=" + orderDate
+                + ", user=" + user
+                + '}';
```
petro-ushchuk, line 76: remove lazy fields from toString

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+    @Inject
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order(
+                shoppingCart.getTickets(), LocalDateTime.now(), shoppingCart.getUser());
+        orderDao.add(order);
```
petro-ushchuk, line 25: use setters instead of constructor.
also, you should put the copy of tickets by creating new ArrayList, instead of putting the original one

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+@Service
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order(
```
petro-ushchuk, line 23: Add checking for `null` for shopping cart. And tickets should not be empty.

petro-ushchuk (changes requested):

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -20,11 +22,16 @@ public class OrderServiceImpl implements OrderService {
 
     @Override
     public Order completeOrder(ShoppingCart shoppingCart) {
-        Order order = new Order(
-                shoppingCart.getTickets(), LocalDateTime.now(), shoppingCart.getUser());
-        orderDao.add(order);
-        shoppingCartService.clearShoppingCart(shoppingCart);
-        return order;
+        if (shoppingCart != null && !shoppingCart.getTickets().isEmpty()) {
```
petro-ushchuk, line 25: what if tickets object is null?

petro-ushchuk (changes requested):

petro-ushchuk (approved):
Good work 🔥

PR #531, opened 2023-04-24
Add order

AleksandrLts:
add Order model
add OrderDao, OrderService and implement them
test all added methods in the main class

### DIFF
--- a/src/main/java/mate/academy/Main.java
+++ b/src/main/java/mate/academy/Main.java
@@ -2,17 +2,39 @@ package mate.academy;
 
 import java.time.LocalDate;
 import java.time.LocalDateTime;
+import mate.academy.exception.AuthenticationException;
+import mate.academy.exception.RegistrationException;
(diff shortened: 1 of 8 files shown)


src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +65,35 @@ public static void main(String[] args) {
...
+        String password = "qwerty";
+        try {
+            authenticationService.register(email, password);
+        } catch (RegistrationException e) {
+            throw new RuntimeException("This user cannot be registered " + email, e);
+        }
+        User user;
+        try {
+            user = authenticationService.login(email, password);
+        } catch (AuthenticationException e) {
```
redmanO-o, line 85: You can use one try block for both operations

nacenik (changes requested):
Good solution! Let’s improve it 😉

nacenik (approved):
LGTM! 🚀

PR #542, opened 2023-05-14
Hw hibernate order

Iskamele:


Iskamele:
Do we have to display all entity information in the getOrdersHistory method?

They were especially considering that tickets are bought in the names of logged-in users.

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+import mate.academy.lib.Service;
+import mate.academy.model.Order;
+import mate.academy.model.ShoppingCart;
+import mate.academy.model.User;
+import mate.academy.service.OrderService;
+import mate.academy.service.ShoppingCartService;
+
+@Service
+public class OrderServiceImpl implements OrderService {
+    private static final LocalDateTime CURRENT_DATE_TIME = LocalDateTime.now();
```
Rommelua, line 16: How many times this constant will be initialized?
Will different orders have different times?

Rommelua (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

Rommelua (approved):

PR #567, opened 2023-07-01
done

aTiukhtii:


src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,52 @@
...
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("from Order o "
+                    + "left join fetch o.user u "
```
Ivan95kos, line 42: It is better to write all SQL commands in upper case

Ivan95kos (approved):

PR #576, opened 2023-07-10
implemented task

chrisnyrka:


src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+@Service
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
```
maxline, line 23: you have constructor with parameter for Order, so you can use it to simplify the code.

maxline (changes requested):
Good job! Let's improve the solution

maxline (approved):
Well done! Approved

PR #579, opened 2023-07-11
HW-Hibernate-11OfJuly

ViktoriiaSvidovska:


src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,53 @@
...
+            if (session != null) {
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("from Order o "
```
Ivan95kos, line 40: It is better to write all SQL commands in upper case
```suggestion
            return session.createQuery("FROM Order o "
```

Ivan95kos (approved):

PR #591, opened 2023-07-14
done

vrnvv:


src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            return session.createQuery("From Order o"
+                            + " left join fetch o.tickets t"
+                            + " left join fetch o.user u"
+                            + " left join fetch t.movieSession m"
+                            + " left join fetch m.movie"
+                            + " left join fetch m.cinemaHall"
+                            + " where o.user = :user", Order.class)
```
kozhukhovsky, line 46:
```suggestion
            return session.createQuery("FROM Order o"
                            + " LEFT JOIN FETCH o.tickets t"
                            // fix other pls
                            + " left join fetch o.user u"
                            + " left join fetch t.movieSession m"
                            + " left join fetch m.movie"
                            + " left join fetch m.cinemaHall"
                            + " where o.user = :user", Order.class)
```

it is better to write SQL keywords in the upper case

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,19 @@
...
+
+        <mapping class="mate.academy.model.CinemaHall" />
+        <mapping class="mate.academy.model.MovieSession" />
+        <mapping class="mate.academy.model.Movie" />
+        <mapping class="mate.academy.model.User" />
+        <mapping class="mate.academy.model.ShoppingCart" />
+        <mapping class="mate.academy.model.Ticket" />
+        <mapping class="mate.academy.model.Order"/>
+    </session-factory>
+</hibernate-configuration>
```
kozhukhovsky, line 19:
```suggestion
</hibernate-configuration>

```

kozhukhovsky (approved):

PR #593, opened 2023-07-14
Homework: Hibernate Order

ViktorBykh:


src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("FROM Order o "
+                    + "LEFT JOIN FETCH o.tickets t "
+                    + "LEFT JOIN FETCH t.movieSession ms "
+                    + "LEFT JOIN FETCH ms.movie "
+                    + "LEFT JOIN FETCH ms.cinemaHall "
+                    + "LEFT JOIN FETCH o.user "
+                    + "WHERE o.user =:user", Order.class);
+            query.setParameter("user", user);
+            return query.getResultList();
```
Poloutinho, line 49: I do it like this, but i think your solution is also valid!

return session.createQuery("FROM Order o "
                    + "LEFT JOIN FETCH ms.movie "
                    + "LEFT JOIN FETCH o.user u "
                    + "WHERE u.id =:userId", Order.class)
                    .setParameter("userId", user.getId())
                    .getResultList();

Poloutinho (approved):

Poloutinho (approved):
Good job keep pushing!

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setTickets(shoppingCart.getTickets());
+        order.setUser(shoppingCart.getUser());
+        orderDao.add(order);
+        shoppingCartService.clearShoppingCart(shoppingCart);
+        return order;
```
sarakhmen, line 27: don't you forget also to fill orderDate field?

sarakhmen (changes requested):
let's slightly improve your solution :)

sarakhmen (approved):
LGTM!

PR #601, opened 2023-07-14
added OrderDao & OrderService

m-pnsk:


AnastasiiaVariuchuk (approved):
good job!

DV7Art (approved):
Good like

src/main/java/mate/academy/model/Ticket.java
```diff
@@ -47,11 +55,12 @@ public void setMovieSession(MovieSession movieSession) {
 
     @Override
     public String toString() {
-        return "Ticket{"
-            + "id=" + id
-            + ", movieSession=" + movieSession
-            + ", user=" + user
-            + '}';
+        return System.lineSeparator()
```
maxline, line 58: It is not good idea to add lineSeparator in `toString()` here . You can use `toString()` later for cases where no need start from the new line.

maxline (approved):
Well done! Approved with minor remark. 
Also want to say that it is good that you are trying to optimize the solution and implement additional  logic that was not necessary for this task. But it is better to push more usual solution for the task review. You can test some additional investigation in separate branch without including into PR.

PR #604, opened 2023-07-17
Added task realization

pkobevko:


Poloutinho (approved):
Overall good job, but i think you choose wrong branch, it should be something like "HW-solution", not "master"

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,83 @@
...
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    @JoinTable(name = "orders_tickets",
+            joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
```
sarakhmen, line 26: redundant annotation

sarakhmen (changes requested):
Good job, but let's slightly improve your solution :) Also, pay attention to the branch you have created PR from (it shouldn't be a master branch)

src/main/java/mate/academy/model/Order.java
pkobevko: without this annotation, Hibernate will create a name for the table itself, and it does not comply with the standards for table naming in SQL

sarakhmen (approved):
LGTM!

PR #607, opened 2023-07-20
added order implementation

krismva:


src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setTickets(shoppingCart.getTickets());
```
olekskov, line 24: Cheklist item:
When passing tickets from shoppingCart to order — pass the copy of this list(you can use ArrayList constructor for this); The reason is that Hibernate will automatically delete these tickets from shoppingCart(in DB) after noticing them in our order object. Let’s do this manually using clear() method from ShoppingCartService after the order with tickets has been created.
https://mate-academy.github.io/jv-program-common-mistakes/hibernate/add-order/add-order-hw

olekskov (changes requested):

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,53 @@
...
+            if (session != null) {
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("from Order o "
```
Ivan95kos, line 41: It is better to write all SQL commands in upper case
```suggestion
            Query<Order> query = session.createQuery("FROM Order o "
```

Ivan95kos (approved):

PR #623, opened 2023-08-18
hibernate-order-hw. Pochtalon

pochtalon:
Created Order model class;
Created interfaces OrderService and OrderDao;
Created classes OrderServiceImpl and OrderDaoImpl;


Ivan95kos (approved):
Great job

PR #629, opened 2023-08-21
implemented OrderService

vaniii1:


src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Override
+    public List<Order> getOrdersHistory(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("from Order o "
+                    + "left join fetch o.tickets t "
+                    + "left join fetch t.movieSession ms "
+                    + "left join fetch ms.movie "
+                    + "left join fetch ms.cinemaHall "
+                    + "where o.user = :user", Order.class);
```
olekskov, line 46:
```suggestion
            Query<Order> query = session.createQuery("FROM Order o "
                    + "LEFT JOIN FETCH o.tickets t "
                    + "LEFT JOIN FETCH t.movieSession ms "
                    + "LEFT JOIN FETCH ms.movie "
                    + "LEFT JOIN FETCH ms.cinemaHall "
                    + "WHERE o.user = :user", Order.class);
```

olekskov (approved):
Good job, approved.
Please consider comment below for further tasks.
Queries are usually more readable when SQL keywords are uppercased

PR #650, opened 2023-09-02
Create models Order. Create DAO ad Service layer for Order

Yulia-Kas:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,61 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @OneToMany
+    private List<Ticket> tickets;
+
+    private LocalDateTime orderDate;
+
+    @ManyToOne
```
aaananas, line 23: eager is a bad practice

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,61 @@
...
+@Entity
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @OneToMany
+    private List<Ticket> tickets;
+
+    private LocalDateTime orderDate;
```
aaananas, line 21: lets add an annotation Column, specify a name

aaananas (changes requested):

PR #673, opened 2023-10-03
implemented orderdao and orderservice

mmosii:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,66 @@
...
+import java.time.LocalDateTime;
+import java.util.List;
+
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
```
ystankevych, line 20:
```suggestion
    @OneToMany
    @JoinTable(name = "orders_tickets",
            joinColumns = @JoinColumn(name = "order_id"),
            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
```
It’s better to make a description for the orders_tickets table with the help of JoinTable annotation. Just look at this table and you will see that the column names are in different case

ystankevych (approved):
Hi! Great job! I approved but left one comment

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Override
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("from Order o "
+                    + "left join fetch o.tickets t "
+                    + "left join fetch t.movieSession ms "
+                    + "left join fetch ms.movie "
+                    + "left join fetch ms.cinemaHall "
+                    + "left join fetch o.user"
```
liliia-ponomarenko, line 46: why left join? can be a movie session without a movie, etc.?

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "Order{"
+                + "id=" + id
+                + ", tickets=" + tickets
+                + ", orderDate=" + orderDate
+                + ", user=" + user + '}';
+    }
```
liliia-ponomarenko, line 70: Better don’t override toString here. it can occur to LIE

liliia-ponomarenko (changes requested):
Great! A few minor mistakes ;)

liliia-ponomarenko (approved):
Great 😊

PR #676, opened 2023-10-04
implemented orders

teract10s:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,72 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    @JoinTable(name = "orders_tickets",
+            joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    private List<Ticket> tickets;
+    private LocalDateTime orderDate;
+    @ManyToOne
```
Andriy-dz, line 27: why EAGER?)

Andriy-dz (approved):

src/main/java/mate/academy/model/Order.java
teract10s: Oh… I just forgot to change from default

PR #714, opened 2023-12-30
Implemented order service:

GoldHolub:
- created model Order
- created OrderDao
- created OrderService

### DIFF
--- a/src/main/java/mate/academy/service/impl/OrderServiceImpl.java
+++ b/src/main/java/mate/academy/service/impl/OrderServiceImpl.java
@@ -1,0 +1,51 @@
+package mate.academy.service.impl;
+
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.List;
+import mate.academy.dao.OrderDao;
+import mate.academy.exception.DataProcessingException;
+import mate.academy.lib.Inject;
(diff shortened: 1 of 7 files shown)


src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+            return order;
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't complete order by ShoppingCart: "
+                                                + shoppingCart, e);
+        }
+    }
+
+    @Override
+    public List<Order> getOrdersHistory(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
```
Rommelua, line 45: what is it for???

Rommelua (changes requested):

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
GoldHolub: For nothing))
My fault. I just added an extra session and try/catch block, even though they already exist at the DAO level.

olekskov (approved):

PR #724, opened 2024-01-15
order class and its dao and service layers created

ivanlavrinenko:


src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        ShoppingCart managedShoppingCart = shoppingCartDao.getByUser(shoppingCart.getUser());
+        User user = managedShoppingCart.getUser();
+        List<Ticket> tickets = managedShoppingCart.getTickets();
+        Order order = new Order();
+        order.setUser(user);
+        order.setTickets(tickets);
+        order.setOrderDate(LocalDateTime.now());
+        return orderDao.add(order);
+    }
```
olekskov, line 30: According to task checklist:
When passing tickets from shoppingCart to order - pass the copy of this list(you can use ArrayList constructor for this); The reason is that Hibernate will automatically delete these tickets from shoppingCart(in DB) after noticing them in our order object. Let's do this manually using clear() method from ShoppingCartService after the order with tickets has been created.

olekskov (changes requested):

olekskov (approved):

PR #732, opened 2024-02-01
solution of hibernate order

furthernull:


src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,31 @@
...
+    private OrderDao orderDao;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setTickets(shoppingCart.getTickets());
+        order.setOrderDate(LocalDateTime.now());
+        order.setUser(shoppingCart.getUser());
+        return orderDao.save(order);
+    }
```
olekskov, line 25: Checklist item:
When passing tickets from shoppingCart to order - pass the copy of this list(you can use ArrayList constructor for this); The reason is that Hibernate will automatically delete these tickets from shoppingCart(in DB) after noticing them in our order object. Let's do this manually using clear() method from ShoppingCartService after the order with tickets has been created.

olekskov (changes requested):

kozhukhovsky (approved):

PR #765, opened 2024-04-08
Implement order.

makaroshyna:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,74 @@
...
+import java.time.LocalDateTime;
+import java.util.List;
+import org.hibernate.annotations.CreationTimestamp;
+
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private long id;
```
DmytroHadiuchko, line 21: I think we have to use wrapper (long -> Long)

DmytroHadiuchko (changes requested):
Good job)  It looks good))
I left only one comment

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,74 @@
...
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    @JoinTable(name = "orders_tickets",
+            joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
```
andrii-hoienko, line 25: Why it is required to create a separate table to implement OneToMany relation? In this case no reason to create separate table I suggest to use joinColumn

andrii-hoienko (changes requested):
Good job! Left some comments

PR #768, opened 2024-04-11
Hibernate order hw solution

vikkushnir:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,68 @@
...
+
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToOne
+    @JoinColumn(name = "user_id")
+    private User user;
```
ValentynSharshon, line 22: Please, check Model structure in the README.md file. Relationship between the Orders and Users tables should be Many to one.

ValentynSharshon (changes requested):
Great job, but need to feex the code a bit to fit the task.

src/main/java/mate/academy/model/Order.java
vikkushnir: Thanks, missing this

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,68 @@
...
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @ManyToOne
+    @JoinColumn(name = "user_id")
+    private User user;
+    @OneToMany
+    private List<Ticket> tickets;
```
karma-o, line 24: Let's specify `@JoinColumn` here as well

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public LocalDateTime getOrderTime() {
+        return orderTime;
+    }
+
+    public void setOrderTime(LocalDateTime orderTime) {
+        this.orderTime = orderTime;
+    }
+
```
karma-o, line 58: equals and hashcode would always be nice to have, even if you are not using them yet

karma-o (approved):
Left some minor comments, GJ!

src/main/java/mate/academy/model/Order.java
vikkushnir: roger that

src/main/java/mate/academy/model/Order.java
vikkushnir: will do

PR #772, opened 2024-04-13
Order implemented

VolandevlodD:
- Created Order model
-Created Order dao
-Created Order service
-Tested in main method

### DIFF
--- a/src/main/java/mate/academy/dao/OrderDao.java
+++ b/src/main/java/mate/academy/dao/OrderDao.java
@@ -1,0 +1,12 @@
+package mate.academy.dao;
+
+import java.util.List;
+import mate.academy.model.Order;
+import mate.academy.model.User;
+
+public interface OrderDao {
+
+    Order add(Order order);
+
+    List<Order> getOrdersHistory(User user);
+}
--- a/src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
+++ b/src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
@@ -1,0 +1,54 @@
+package mate.academy.dao.impl;
+
+import java.util.List;
+import mate.academy.dao.OrderDao;
+import mate.academy.exception.DataProcessingException;
+import mate.academy.lib.Dao;
+import mate.academy.model.Order;
+import mate.academy.model.User;
+import mate.academy.util.HibernateUtil;
+import org.hibernate.Session;
+import org.hibernate.SessionFactory;
+import org.hibernate.Transaction;
+import org.hibernate.query.Query;
+
+@Dao
+public class OrderDaoImpl implements OrderDao {
+    private final SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
+
+    @Override
+    public Order add(Order order) {
+        Session session = null;
+        Transaction transaction = null;
+        try {
+            session = sessionFactory.openSession();
+            transaction = session.beginTransaction();
+            session.persist(order);
+            transaction.commit();
+            return order;
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can`t add order" + order, e);
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public List<Order> getOrdersHistory(User user) {
+        try (Session session = sessionFactory.openSession()) {
+            String getOrdersByUserQuery = "FROM mate.academy.model.Order o JOIN FETCH o.tickets t "
+                    + "JOIN FETCH t.movieSession ms JOIN FETCH ms.movie "
+                    + "JOIN FETCH ms.cinemaHall WHERE o.user = :user";
+            Query<Order> query = session.createQuery(getOrdersByUserQuery, Order.class);
+            query.setParameter("user", user);
+            return query.getResultList();
+        } catch (Exception e) {
+            throw new DataProcessingException("Can`t get orders history for user: " + user, e);
+        }
+    }
+}
--- a/src/main/java/mate/academy/model/Order.java
+++ b/src/main/java/mate/academy/model/Order.java
@@ -1,0 +1,64 @@
+package mate.academy.model;
+
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.JoinColumn;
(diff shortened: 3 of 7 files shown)


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,64 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    @JoinTable(name = "orders_tickets", joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
+    private List<Ticket> tickets;
+    private LocalDateTime orderDate;
+    @ManyToOne
+    @JoinColumn(name = "user_id")
```
nklimovych, line 25: In my opinion this line is redundant, but it's up to you)

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        List<Ticket> shoppingCartTickets = shoppingCart.getTickets();
+        Order order = new Order();
+        order.setTickets(new ArrayList<Ticket>(shoppingCartTickets));
+        shoppingCartService.clearShoppingCart(shoppingCart);
+        order.setUser(shoppingCart.getUser());
+        order.setOrderDate(LocalDateTime.now());
+        orderDao.add(order);
```
nklimovych, line 31: This method looks a bit over complicated. It would be better to add empty lines or refactor it.

src/main/java/mate/academy/dao/OrderDao.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.List;
+import mate.academy.model.Order;
+import mate.academy.model.User;
+
+public interface OrderDao {
+
+    Order add(Order order);
+
+    List<Order> getOrdersHistory(User user);
```
nklimovych, line 11: According to README.MD:

- To get the order for a specific user create the method getByUser().

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+            if (session != null) {
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public List<Order> getOrdersHistory(User user) {
+        try (Session session = sessionFactory.openSession()) {
+            String getOrdersByUserQuery = "FROM mate.academy.model.Order o JOIN FETCH o.tickets t "
```
nklimovych, line 44: Why not use import instead of hardcoding it in the query?

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+                    + "JOIN FETCH t.movieSession ms JOIN FETCH ms.movie "
+                    + "JOIN FETCH ms.cinemaHall WHERE o.user = :user";
```
nklimovych, line 46: I believe we should search for matches where o.user.id = :id.

nklimovych (changes requested):
Looks great, left you some suggestions

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
VolandevlodD: It's not a matter of using 'import'; I simply renamed my table from 'Order' to 'orders.' That's why I'm using the full name here.

karma-o (changes requested):
left several comments

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.List;
+
+@Entity(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    @JoinTable(name = "orders_tickets", joinColumns = @JoinColumn(name = "order_id"),
+            inverseJoinColumns = @JoinColumn(name = "ticket_id"))
```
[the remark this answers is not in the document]
VolandevlodD: Thank you! I employed it solely to ensure that column names in the database maintain a consistent singular form. For instance, without this annotation, names like 'order_id' and 'ticket_id' would adopt a plural form when created by Hibernate, whereas every other table in my database follows a singular naming convention.
Is it ok to use this annotation for this purpose?
image

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,63 @@
...
+    private List<Ticket> tickets;
+    private LocalDateTime orderDate;
+    @ManyToOne
+    private User user;
```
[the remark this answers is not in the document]
VolandevlodD: I've tested Hibernate in main, and it automatically creates a column named 'user_id'.
image

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
VolandevlodD: What I mean is that I've changed the column name from 'order' to 'orders' as described in the README.
image
image
So, now I need to either use 'orders' as the name or the full path here; otherwise, I'll encounter an error.
image
image
That is why I used the full path qualifier, but it also works when I use name orders for the table:
image

src/main/java/mate/academy/model/Order.java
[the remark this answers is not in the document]
andrii-hoienko: You don`t need JoinTable because to create one to many relation it is not necessary to create a separate table, if you want to specify custom filed names I suggest using joinColumn

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,65 @@
...
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.JoinTable;
+import jakarta.persistence.ManyToOne;
+import jakarta.persistence.OneToMany;
+import java.time.LocalDateTime;
+import java.util.List;
+
+@Entity(name = "orders")
```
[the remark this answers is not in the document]
VolandevlodD: Got it, thanks for clarifying. Previously, I was renaming the entity, but now it's all sorted out.

andrii-hoienko (approved):
Good job!

PR #773, opened 2024-04-13
Added implementation for task hibernate order hw. 

nklimovych:
Created Order entity, Dao and Service layers. Also hibernate the configuration file.

### DIFF
--- a/src/main/java/mate/academy/Main.java
+++ b/src/main/java/mate/academy/Main.java
@@ -2,16 +2,25 @@ package mate.academy;
 
 import java.time.LocalDate;
 import java.time.LocalDateTime;
+import mate.academy.exception.RegistrationException;
+import mate.academy.lib.Injector;
 import mate.academy.model.CinemaHall;
 import mate.academy.model.Movie;
 import mate.academy.model.MovieSession;
+import mate.academy.model.ShoppingCart;
+import mate.academy.model.User;
+import mate.academy.security.AuthenticationService;
 import mate.academy.service.CinemaHallService;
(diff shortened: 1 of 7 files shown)


src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +54,33 @@ public static void main(String[] args) {
...
+
+        AuthenticationService authenticationService =
+                (AuthenticationService) injector.getInstance(AuthenticationService.class);
+        User newUser = null;
+        try {
+            newUser = authenticationService.register("newuser@gmail.com", "pass");
+            System.out.println(newUser);
+        } catch (RegistrationException e) {
+            throw new RuntimeException("Can't register a user: " + newUser, e);
+        }
```
karma-o, line 74: in main we can rethrow exception to the method signature instead of try/catch but that is up to you

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,76 @@
...
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany(fetch = FetchType.LAZY)
+    private List<Ticket> tickets;
+    private LocalDateTime orderDate;
+    @ManyToOne(fetch = FetchType.LAZY)
+    private User user;
```
karma-o, line 24: let's specify the join column to not let hibernate create a third table for each relation

karma-o (approved):
Commented a couple of minor things, please check them out

PR #774, opened 2024-04-16
added 'Order' model, DAO & Service

DanilKi:


src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+import org.hibernate.query.Query;
+
+@Dao
+public class OrderDaoImpl implements OrderDao {
+    @Override
+    public Order add(Order order) {
+        Session session = null;
+        Transaction transaction = null;
+        try {
+            session = HibernateUtil.getSessionFactory().openSession();
```
MAMentorFX, line 21: Why can't we getSessionFactory only once for this class?

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+    @Inject
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setUser(shoppingCart.getUser());
+        order.setTickets(new ArrayList<>(shoppingCart.getTickets()));
```
MAMentorFX, line 26:
```suggestion
        order.setTickets(List.of(shoppingCart.getTickets()));
```

MAMentorFX (approved):

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
DanilKi: You're right, but checklist.md file should also be corrected then

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
DanilKi: Yes, of course we can. I did so only because it was done in such a way in other DAO's by the author of this project.

PR #778, opened 2024-04-19
hibernate-order-hw-solution 1

DojiDoe:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,65 @@
...
+import jakarta.persistence.Id;
+import jakarta.persistence.ManyToOne;
+import jakarta.persistence.OneToMany;
+import jakarta.persistence.Table;
+import java.time.LocalDateTime;
+import java.util.List;
+
+@Entity
+@Table(name = "orders")
+public class Order {
```
MAMentorFX, line 15: Suggest @Getter @Setter to remove all boilerplate code

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+
+@Service
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
```
MAMentorFX, line 23: @Transactional. You make more than one call to DB

MAMentorFX (changes requested):

okuzan (approved):

PR #784, opened 2024-05-20
implemented DAO and Service for Order model

nastasiiaKulinich:


src/main/java/mate/academy/dao/OrderDao.java
```diff
@@ -0,0 +1,16 @@
+package mate.academy.dao;
+
+import java.util.List;
+import java.util.Optional;
+import mate.academy.model.Order;
+import mate.academy.model.User;
+
+public interface OrderDao {
```
Elena-Bruyako, line 8: Minor comment, you actually use only add and getByUser methods, so it's not necessary to implement 
`Optional<Order> getById(Long id) and List<Order> getAll() `methods

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #788, opened 2024-05-26
Order impl

petro2061:
Added implementation Order entity class. Order Dao and Order Sevise also implemented. In Main method added test for new methods.

### DIFF
--- a/src/main/java/mate/academy/model/Order.java
+++ b/src/main/java/mate/academy/model/Order.java
@@ -1,0 +1,71 @@
+package mate.academy.model;
+
+import jakarta.persistence.Entity;
+import jakarta.persistence.FetchType;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.JoinTable;
+import jakarta.persistence.ManyToOne;
+import jakarta.persistence.OneToMany;
+import jakarta.persistence.Table;
+import java.time.LocalDateTime;
+import java.util.List;
+
+@Entity
+@Table (name = "orders")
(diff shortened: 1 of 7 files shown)


src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order newOrderBaseShoppingCart = new Order();
+        newOrderBaseShoppingCart.setOrderDate(LocalDateTime.now());
+        newOrderBaseShoppingCart.setUser(shoppingCart.getUser());
+        List<Ticket> ticketsFromShoppingCart = shoppingCart.getTickets();
+        newOrderBaseShoppingCart.setTickets(ticketsFromShoppingCart);
```
okuzan, line 28: checklist issue 
image

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.List;
+
+@Entity
+@Table (name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue (strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    @JoinTable(name = "orders_tickets",
```
okuzan, line 23: join table is typically used for many to many relationship, not one to many

okuzan (changes requested):

petro2061:
fixed implementation method completeOrder in OrderServiseImpl class and changed annotation JoinTable on JoinColumn in Order class

MentorMateAcademy (approved):

PR #796, opened 2024-06-18
resolve order task

YouraSoliar:


src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService shoppingCartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order(shoppingCart.getUser(), shoppingCart.getTickets(),
+                shoppingCart.getTickets().get(0).getMovieSession().getShowTime());
```
sarakhmen, line 23: why do you use moviesession showtime as an order date?

sarakhmen (changes requested):
good job, but let's slightly improve your solution :)

Elena-Bruyako (approved):

PR #802, opened 2024-07-09
MateAcademy/hibernate-order-hw - Bartosz Wójcik solution

banderan:


banderan:
I wanted to try a solution using validate however after a few hours I gave up and I don't know what I did wrong in those .yml files

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+public class OrderServiceImpl implements OrderService {
+    @Inject
+    private OrderDao orderDao;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        User user = shoppingCart.getUser();
+        List<Ticket> tickets = shoppingCart.getTickets();
+        LocalDateTime now = LocalDateTime.now();
+        Order order = new Order();
```
beatarjn, line 24: I would advise to extract Order creation to a separate method e.g createOrder() so it is more readable and modular.

src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +55,34 @@ public static void main(String[] args) {
...
+            throw new RuntimeException(e);
+        }
+
+        ShoppingCartService shoppingCartService =
+                (ShoppingCartService) injector.getInstance(ShoppingCartService.class);
+        ShoppingCart byUser = shoppingCartService.getByUser(registered);
+
+        OrderService orderService =
+                (OrderService) injector.getInstance(OrderService.class);
+        orderService.completeOrder(byUser);
```
beatarjn, line 85: You are only calling methods from OrderService and not using a result. How about adding some methods that would check if the called methods behave correctly? For example assign orderService.completeOrder(byUser) to a variable and then do even a simple check if it's not null or print it. The same for method getOrdersHistory() - check if list is not null or print it.

src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +55,34 @@ public static void main(String[] args) {
...
         movieSessionService.add(tomorrowMovieSession);
         movieSessionService.add(yesterdayMovieSession);
 
         System.out.println(movieSessionService.get(yesterdayMovieSession.getId()));
         System.out.println(movieSessionService.findAvailableSessions(
                         fastAndFurious.getId(), LocalDate.now()));
+
+        AuthenticationService authenticationService =
+                (AuthenticationService) injector.getInstance(AuthenticationService.class);
+        User user = new User();
```
beatarjn, line 69: Extracting the user creation into a separate method can improve readability and modularity.

src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,66 @@
...
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    private LocalDateTime orderDate;
+    @OneToMany
+    private List<Ticket> tickets;
+    @ManyToOne
```
beatarjn, line 22: By default, @ManyToOne relationships use FetchType.EAGER, which means that the associated entity is loaded immediately along with the owning entity. However, it is preferable to use FetchType.LAZY to improve performance by loading the associated entity only when it is accessed. 
You can find more info here: https://thorben-janssen.com/common-hibernate-mistakes-cripple-performance/#Mistake_1_Use_Eager_Fetching

beatarjn (changes requested):
The problem with "not working validate method" would be a good point to be discussed on the Q&A session.

banderan:
I forgot  about Q&A but in next week i ll ask mentors. Thanks

src/main/java/mate/academy/service/impl/OrderServiceImpl.java
MateuszRostek, line 1: After order is completed, you should clear the shopping cart (Using ShoppingCartService and injection).

MateuszRostek (approved):
Good job, see minor advice.

beatarjn (approved):
Changes implemented, well done

nowickimj:
> I wanted to try a solution using validate however after a few hours I gave up and I don't know what I did wrong in those .yml files

Not sure why would you like to use `validate` property here - it won't apply any changes to your database, hence the exception informing you on missing tables.
`create-drop` is perfect your case. 

Btw not sure about liquibase configuration and changelog definition, since you didn't put any dependency to actually use it.

PR #805, opened 2024-07-17
hibernate-order-hw

YanaBN:
Created Order model, then created its DAO and implemented it, used it in OrderServiceImpl, used main() to check that everything is working, and ran mvn clean package afterward to make sure there are no mistakes in the code;

### DIFF
--- a/src/main/java/mate/academy/Main.java
+++ b/src/main/java/mate/academy/Main.java
@@ -2,16 +2,25 @@ package mate.academy;
 
 import java.time.LocalDate;
 import java.time.LocalDateTime;
+import mate.academy.lib.Injector;
 import mate.academy.model.CinemaHall;
 import mate.academy.model.Movie;
 import mate.academy.model.MovieSession;
+import mate.academy.model.User;
 import mate.academy.service.CinemaHallService;
 import mate.academy.service.MovieService;
 import mate.academy.service.MovieSessionService;
+import mate.academy.service.OrderService;
+import mate.academy.service.ShoppingCartService;
+import mate.academy.service.UserService;
 
 public class Main {
+    private static final Injector injector =
+            Injector.getInstance("mate.academy");
+
     public static void main(String[] args) {
-        MovieService movieService = null;
+        MovieService movieService =
+                (MovieService) injector.getInstance(MovieService.class);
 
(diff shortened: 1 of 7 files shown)


src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +54,42 @@ public static void main(String[] args) {
...
+        MovieSessionService movieSessionService =
+                (MovieSessionService) injector.getInstance(MovieSessionService.class);
         movieSessionService.add(tomorrowMovieSession);
         movieSessionService.add(yesterdayMovieSession);
 
         System.out.println(movieSessionService.get(yesterdayMovieSession.getId()));
         System.out.println(movieSessionService.findAvailableSessions(
                         fastAndFurious.getId(), LocalDate.now()));
+
+        User userA = new User();
```
Elena-Bruyako, line 66: bad naming

src/main/java/mate/academy/Main.java
```diff
@@ -44,12 +54,42 @@ public static void main(String[] args) {
...
+        userA.setEmail("Alice@gmail.com");
+        userA.setPassword("Alice1234");
+        User userB = new User();
```
Elena-Bruyako, line 69: same

src/main/java/mate/academy/dao/impl/OrderDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+                session.close();
+            }
+        }
+    }
+
+    @Override
+    public List<Order> getByUser(User user) {
+        try (Session session = HibernateUtil.getSessionFactory().openSession()) {
+            Query<Order> query = session.createQuery("FROM Order o "
+                    + "LEFT JOIN FETCH o.user "
```
Elena-Bruyako, line 42:
```suggestion
                + "LEFT JOIN FETCH o.tickets t "
                + "JOIN FETCH o.user u "
                + "JOIN FETCH t.movieSession ms "
                + "JOIN FETCH ms.movie m "
                + "JOIN FETCH ms.cinemaHall c "
                + "WHERE o.user = :user"
```

Elena-Bruyako (changes requested):

PR #816, opened 2024-08-15
hibernate-order-hw solution

c-kalenyk:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.List;
+
+@Entity
+@Table(name = "orders")
+public class Order {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @OneToMany
+    @JoinTable(name = "orders_tickets",
```
sarakhmen, line 23: why did you use jointable here?

sarakhmen (changes requested):
let's lightly improve your solution :)

PR #823, opened 2024-08-29
hibernate-order-hw task solution

SvNadorozhnyi:


src/main/java/mate/academy/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+    private OrderDao orderDao;
+    @Inject
+    private ShoppingCartService cartService;
+
+    @Override
+    public Order completeOrder(ShoppingCart shoppingCart) {
+        Order order = new Order();
+        order.setOrderDate(LocalDateTime.now());
+        order.setUser(shoppingCart.getUser());
+        order.setTickets(shoppingCart.getTickets());
```
Elena-Bruyako, line 26: When passing tickets from shoppingCart to order - pass the copy of this list(you can use ArrayList constructor for this); The reason is that Hibernate will automatically delete these tickets from shoppingCart(in DB) after noticing them in our order object. Let's do this manually using clear() method from ShoppingCartService after the order with tickets has been created.
```suggestion
        order.setTickets(new ArrayList<>(shoppingCart.getTickets()));
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #843, opened 2024-11-29
added model Order to the Cinema project

GarikSlavsky:


src/main/java/mate/academy/model/Order.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Column(name = "order_date")
+    private LocalDateTime orderDate;
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "user_id")
+    private User user;
+
+    public Order() {
+    }
```
ahoienko, line 38: Redundant constructor, please left only methods that you use

